PR #13, opened 2016-01-02
butterknife / jcenter

Rainer-Lang:
Butterknife can be used


Rainer-Lang:
not good

mikepenz:
was machst du? :D

PR #77, opened 2016-03-17
Slight optimisation when we know the size of list

shubhamchaudhary:


mikepenz:
@shubhamchaudhary thanks. This is a really good one ;)

mattiasbe:
**Suggestion:** Collections.emptyList(); Instead to avoid allocation altogether.
**Interesting note:** At least in regular java (since java 1.7update 40), `ArrayList `(and `HashMap`) where nothing has been added yet doesn't consume "any" memory, nothing is preallocated for the storage.
Meaning `new ArrayList<>()` or `new ArrayList<>(0)` doesn't matter for memory usage.

PR #230, opened 2016-09-30
Moved the custom drag handle setup to a utility class

MFlisar:
must now be called manually in bindView of the IExtendedDraggable item


mikepenz:
@MFlisar I think so we should also move the interface for it?

PR #362, opened 2017-03-09
Fix: failed withSavedInstanceState if withPositionBasedStateManagement(false)

fikr4n:
I can't think of better idea to solve #360 other than this one-line fix, or to loop until exception thrown at `item = getItem(i);` instead of `i < size` (python style).


fikr4n:
Because FastAdapter named "fast" so I think about performance 😄

PR #425, opened 2017-05-22
Drop callback support + docs added

MFlisar:
* added some docs to the `CustomEventHook`
* extended the `ItemTouchCallback` and added an `itemTouchDropped` event
* extended the `SimpleDragCallback` to fire the drop event

drop callback is based on this issue: https://github.com/mikepenz/FastAdapter/issues/59

This change is not backwards compatible, every instance of `ItemTouchCallback` must implement now the `itemTouchDropped(int oldPosition, int newPosition)` function...


library-core/src/main/java/com/mikepenz/fastadapter/listeners/EventHook.java
```diff
@@ -10,9 +10,20 @@
 import java.util.List;
 
 public interface EventHook<Item extends IItem> {
+
+    /*
+	 * return the view for this hook that the listener should be bound to
```
FabianTerhorst: Doc should start with upper case

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
```diff
@@ -21,6 +21,9 @@
     private ItemTouchCallback mCallbackItemTouch; // interface
     private boolean mIsDragEnabled = true;
 
+    private Integer mFrom = null;
+    private Integer mTo = null;
```
FabianTerhorst: there are null by default

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
```diff
@@ -63,6 +66,17 @@ public boolean isLongPressDragEnabled() {
...
+        // remember the from/to positions
+        if (target.itemView.getTag() instanceof IDraggable) {
+            if (((IDraggable) target.itemView.getTag()).isDraggable()) {
+                if (mFrom == null) {
+                    mFrom = viewHolder.getAdapterPosition();
+                }
+                mTo = target.getAdapterPosition();
+            }
+        }
+
```
FabianTerhorst: remove space

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
```diff
@@ -63,6 +66,17 @@ public boolean isLongPressDragEnabled() {
 
     @Override
     public boolean onMove(RecyclerView recyclerView, RecyclerView.ViewHolder viewHolder, RecyclerView.ViewHolder target) {
+
```
FabianTerhorst: remove space

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
```diff
@@ -97,4 +111,18 @@ public int getDragDirs(RecyclerView recyclerView, RecyclerView.ViewHolder viewHo
     public void onSwiped(RecyclerView.ViewHolder viewHolder, int direction) {
         // swiped disabled
     }
+
+    @Override
+    public void clearView(RecyclerView recyclerView, RecyclerView.ViewHolder viewHolder) {
+        super.clearView(recyclerView, viewHolder);
+
```
FabianTerhorst: remove space

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
```diff
@@ -63,6 +66,17 @@ public boolean isLongPressDragEnabled() {
 
     @Override
     public boolean onMove(RecyclerView recyclerView, RecyclerView.ViewHolder viewHolder, RecyclerView.ViewHolder target) {
+
+        // remember the from/to positions
+        if (target.itemView.getTag() instanceof IDraggable) {
+            if (((IDraggable) target.itemView.getTag()).isDraggable()) {
+                if (mFrom == null) {
```
FabianTerhorst: Are you sure this is needed? Because when you start moving a second time from another positon the mFrom is wrong?

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
FabianTerhorst: you could also replace them by int with RecyclerView.NO_POSITION as default

library-extensions/src/main/java/com/mikepenz/fastadapter_extensions/drag/SimpleDragCallback.java
FabianTerhorst: Haven´t tested, just noticed inside the code.

MFlisar: What do you mean? The `mFrom == null` check? This is necessary because we want to save the very first from position only and don't want to overwrite this value while dragging around the item

MFlisar:
Updated the examples.

Additionally I added a method to the `DragDropUtil`. There's a bug in the current examples. If you scroll down a list and than start a drag and drop, move your finger to the top of the screen and let the list scroll up and then move the finger down to the list again, the `itemTouchOnMove´ will get the top most position before the scrolling and the top position. This means, you must swap items one for one in this case...

The utility method would do this and imho should be used in all examples instead of current swapping of from and to position.

FabianTerhorst:
Could you replace the logic inside the sample with the new utils method?

MFlisar:
Sure, just wanted to know if that's desired

FabianTerhorst:
The builds are failing because of bintray´s server issues.

PR #482, opened 2017-07-16
[WIP] [V3] New ModelAdapter

FabianTerhorst:
* Add an IdDistributor interface to make id distribution to each item flexible and extendable. This change also stores the id distributor instance to the item adapter. The default implementation is stored static inside the interface so the atomic long is shared with all item adapters to not collide with an second item adapter instance.
When you want to make an own implementation that depends on an id increasement you have to share the implementation to all item adapters or make the id static.


library-core/src/main/java/com/mikepenz/fastadapter/adapters/ItemAdapter.java
```diff
@@ -10,12 +10,7 @@
 public class ItemAdapter<Item extends IItem> extends ModelAdapter<Item, Item> {
 
     public ItemAdapter() {
-        super(new IInterceptor<Item, Item>() {
-            @Override
-            public Item intercept(Item item) {
-                return item;
-            }
-        });
+        super((IInterceptor<Item, Item>) IInterceptor.DEFAULT);
```
FabianTerhorst: static generic doesnt work, not sure about an other way without an cast.

library-core/src/main/java/com/mikepenz/fastadapter/adapters/ModelAdapter.java
```diff
@@ -61,13 +62,33 @@ private Item intercept(Model model) {
...
+     * @return this
+     */
+    public ModelAdapter<Model, Item> withIdDistributor(IIdDistributor<Item> idDistributor) {
+        this.mIdDistributor = idDistributor;
+        return this;
+    }
+
+    public IIdDistributor<Item> getIdDistributor() {
+        if (mIdDistributor == null) {
+            return (IIdDistributor<Item>) IIdDistributor.DEFAULT;
```
FabianTerhorst: static generic doesnt work, not sure about an other way without an cast.

PR #507, opened 2017-08-05
Fix typo in comment

bernaferrari:


FabianTerhorst:
Not sure if we should accept pr´s for not v3. Could you rebase to the v3_new branch.

FabianTerhorst:
The typo is present in almost all sample activities. Could you fix them all after rebase?

PR #669, opened 2018-04-26
Fix NPE when add previously filled adapter to new FastAdapter instance

bejibx:


FabianTerhorst:
Why not calling it inside addAdapter for the newly added adapter only, instead of running it on all adapters again that are already added?

bejibx:
@FabianTerhorst sorry, I don't understand your idea. Can you please describe it more detailed?

bejibx:
@mikepenz @FabianTerhorst can someone please help me with this pr? I don't quite understand what should I do?

PR #682, opened 2018-05-28
Fix #669

FabianTerhorst:


FabianTerhorst:
This would require to add the internal used mapPossibleTypes to the interface, but im not seeing an problem here because withFastAdapter is also internal only.

library-core/src/main/java/com/mikepenz/fastadapter/FastAdapter.java
```diff
@@ -202,8 +202,10 @@ public void setTypeInstanceCache(ITypeInstanceCache<Item> mTypeInstanceCache) {
      */
     public <A extends IAdapter<Item>> FastAdapter<Item> addAdapter(int index, A adapter) {
         mAdapters.add(index, adapter);
+        adapter.withFastAdapter(this);
+        adapter.mapPossibleTypes(adapter.getAdapterItems());
         for (int i = 0; i < mAdapters.size(); i++) {
-            mAdapters.get(i).withFastAdapter(this).setOrder(i);
```
FabianTerhorst: the other adapters are already added to the fast adapter so there is no need to call the setter multiple times

FabianTerhorst:
This won't work when the item adapter isn't added to any fastadapter. Not sure if we should ever support adding items to an item adapter without an fast adapter.

FabianTerhorst:
Added support for nullable fast adapter, but not sure if we want that.

FabianTerhorst:
@bejibx

bejibx:
Looking great! Can't tell if support adding items without FastAdapter will cause unexpected side-effects. Anyway this should not affect normal usage cases.

PR #730, opened 2018-10-16
[V4] Improved expandable

FabianTerhorst:
Removes ISubItem interface by moving the parent declaration to IExpandable.


FabianTerhorst:
Travis won't be happy, because thats an binary incompatibility with the material drawer.

FabianTerhorst:
TODO:

- move select, deselect handling from expandable extension to select extension

PR #762, opened 2019-03-19
Fix diff utils

FabianTerhorst:


KiminRyu:
I think it would be nice if it is released soon. (expected v4-rc04)

PR #763, opened 2019-03-24
Update README.md

barodapride:


FabianTerhorst:
Please don't allocate a new stringholder on each bind call. Take a look at https://github.com/mikepenz/FastAdapter/blob/feature/v4/library-extensions-ui/src/main/java/com/mikepenz/fastadapter/ui/items/SingleLineItem.kt#L36

barodapride:
I still don't understand what to do. The code in the readme doesn't work. Is SingleLineItem an example I'm supposed to follow?

PR #803, opened 2019-06-30
JvmFields

### LINKED ISSUE #800: Just add @JvmStatic
> As a result of the Kotlin migration the static methods can now be accessed via FastAdapter.Companion. E.g.: FastAdapter.Companion.with(...)

Just add `@JvmStatic` on methods. Refactoring to Kotlin will definitely add breaking changes, but doesn't mean you have to break the things which have a working solution to

AllanWang:
Resolves #800

Most functions are already marked as `@JvmStatic`. I added `@JvmField` to a few of the existing fields though

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -5,7 +5,7 @@
 v4 is a huge release changing most of the codebase to Kotlin. This comes with many refactors, and as a result of that with many breaking API changes. 
 We put a lot of focus on type safety with this release, as such this release is a lot more strict and tries to prevent as many potential bad type mixups as possible.
 
-* As a result of the Kotlin migration the static methods can now be accessed via `FastAdapter.Companion`. E.g.: `FastAdapter.Companion.with(...)`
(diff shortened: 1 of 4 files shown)


MIGRATION.md
```diff
@@ -5,7 +5,7 @@
 v4 is a huge release changing most of the codebase to Kotlin. This comes with many refactors, and as a result of that with many breaking API changes. 
 We put a lot of focus on type safety with this release, as such this release is a lot more strict and tries to prevent as many potential bad type mixups as possible.
 
-* As a result of the Kotlin migration the static methods can now be accessed via `FastAdapter.Companion`. E.g.: `FastAdapter.Companion.with(...)`
+* For compatibility, most existing static fields and functions remain static in Java. For newer functions, accessing from Java may require using the `Companion` class. For instance, `FastAdapter.example()` becomes `FastAdapter.Companion.example()`
```
iamareebjamal: That's wrong, no change is required in Java if @JvmStatic is added. In kotlin, FastAdapter.Companion import is needed

AllanWang: The note is for the future in case we don't add `@JvmStatic`. In that case, the fallback behaviour would be to use `Companion`. To maintain backwards compatibility, we've added static fields/methods to existing functions

PR #819, opened 2019-09-10
Added LayoutContainer in FastAdapter.ViewHolder

HirogaKatageri:
Added LayoutContainer in order to utilise Android Extensions to access and cache views.

Ref: https://proandroiddev.com/kotlin-android-extensions-using-view-binding-the-right-way-707cd0c9e648

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,17 @@ implementation "com.google.android.material:material:${androidX}"
 implementation "com.mikepenz:materialize:${latestVersion}" // at least 1.2.0
 ```
 
+If unable to use Android Extensions to access views in ViewHolder, add line below in app gradle.
+```gradle
+android{
+    ...
+    androidExtensions {
+        experimental = true
+    }
+}
+
+```
+
 ## v4.x.y
 > Major release, migrates fully to Kotlin. Check out the changelog or the [MIGRATION GUIDE](https://github.com/mikepenz/FastAdapter/blob/develop/MIGRATION.md for more details
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,6 @@
 apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-android'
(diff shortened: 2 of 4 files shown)


AllanWang:
I don't think this should be added. Not everyone wants to use kotlin android extensions, and plus it isn't recommended. It is very easy to add it per project, as shown in your PR

HirogaKatageri:
I understand now, you're right. I wasn't aware of the synthetics debate. At first the reason I wanted to include this was because there are many developers that do.

`
itemView.textView.text = "String Here"
`

Which has performance issues.
Compared to adding LayoutContainer which turns above code into

`
textView.text = ""
`

Since Kotlin Android Extensions showed up there are many developers that have joined the band wagon myself included and are unaware of such details.

@AllanWang If this was created as an another module that adds support for Android Extensions that contains another ViewHolder that inherits FastAdapter.ViewHolder and LayoutContainer, would this be more acceptable? That way it is not directly included in the FastAdapter.ViewHolder but is optional for the user to include it?

FabianTerhorst:
As AllanWang correctly already said kotlin extensions aren't recommended by Google and I also don't think they should be added to the core.

AllanWang:
You certainly could, but at that point it may be worth considering whether a new submodule is worth it for the 3 line change. It may also be interesting to see why ViewHolders aren't yet supported out of experimental yet, or whether there are any plans to update the extension at all. Speaking of helpers, I do have my own ViewHolder that is based on LayoutBinding that I plan on sharing eventually, but that's for another PR. While I use it mainly to get ids, I find it a lot nicer to use, and it comes with some benefits beyond a few lines

AllanWang:
I also doubt that 

```kotlin
itemView.textView.text = "String Here"
```

has performance issues if it is bound once at the beginning. If anything it's faster if you know you'll use the ID. In the end, everything uses `findViewById`

HirogaKatageri:
Understood then I'll just close this and wait for your future PR. Thank you for the feedback :)

PR #888, opened 2020-04-30
Suggestion to listen to the drag start event (#886)

RobbWatershed:


library-extensions-drag/src/main/java/com/mikepenz/fastadapter/drag/ItemTouchCallback.kt
```diff
@@ -1,7 +1,16 @@
...
+import androidx.recyclerview.widget.RecyclerView
+
 interface ItemTouchCallback {
 
+    /**
+     * Called when an item enters drag state
+     *
+     * @param viewHolder dragged ViewHolder
+     */
+    fun itemTouchStartDrag(viewHolder : RecyclerView.ViewHolder)
```
mikepenz, line 12: could you perhaps provide a default impl for this one? so people who already use this interface won't have a compile time issue.

mikepenz: This will also fix the PR build failure I believe: 

image

RobbWatershed: Shoot... I should've rebuilt the whole solution including the sample activities, not only the drag module 😅 

I'll do that very soon. Thanks for the reply.

RobbWatershed:
This is the one. I've successfuly implemented and tested the feature in the sample app (see `SwipeListActivity`).

I took the liberty to create `IDraggableViewHolder` in the app itself. Imho, it's a design choice that shouldn't be constrained by the library.

mikepenz:
Thank you so much :)

RobbWatershed:
I just realized there's a small issue with what I'm trying to do : the special "drag highlight" visual is supposed to disappear whenever the item is dropped. It works fine when it actually changes places... but when you start dragging and redrop it at its initial position, `itemTouchDropped` is not called.

The reason for that is you've decided to call it only when the item has changed its position (see https://github.com/mikepenz/FastAdapter/blob/fde5ff985ebf1baed57c97f1329fa53408ce4719/library-extensions-drag/src/main/java/com/mikepenz/fastadapter/drag/SimpleDragCallback.kt#L101)

@mikepenz I'd like to discuss our options here. Would you find it appropriate to have an new optional parameter to set on `SimpleDragCallback` (e.g. `SimpleDragCallback.notifyAllDrops`) that would be off by default (current behaviour) ?

mikepenz:
@RobbWatershed so the simple drag callback was designed to be the most simple variant. we might just want to provide a normal DragCallback instead which will do kind of more like notify in all cases no no matter what. 

I suppose changing it now would break existing implementations potentially as such that's not really an option we have right now I believe?

RobbWatershed:
> I suppose changing it now would break existing implementations potentially as such that's not really an option we have right now I believe?

Not necessarily. I had something like that in mind :
```java
private var notifyAllDrops = false // Default behaviour of v5.0.0

fun setNotifyAllDrops(notifyAllDrops : Boolean) {
    this.notifyAllDrops = notifyAllDrops 
}

 override fun clearView(recyclerView: RecyclerView, viewHolder: RecyclerView.ViewHolder) {
    super.clearView(recyclerView, viewHolder)
    if (notifyAllDrops || (from != RecyclerView.NO_POSITION && to != RecyclerView.NO_POSITION)) {
        callbackItemTouch?.itemTouchDropped(from, to)
    }
    // reset the from/to positions
    to = RecyclerView.NO_POSITION
    from = to
}
```

Forcing users to use a vanilla `ItemTouchHelper.SimpleCallback` to get all drop events would also prevent them from using `SimpleSwipeDragCallback`, which is not very flexible.

RobbWatershed:
Thanks for the merge ! I'm gonna make a new PR for what I suggested regarding the drop event

PR #925, opened 2020-10-08
Set correct default value for toggleExpandable and collapse

Ceees2:
This caused onBind being called while animations are running resulting in weird behaviour of animations.

Docs also suggest this should be false.


mikepenz:
@Ceees2 can you please highlight which weird behaviour of animations you see?

I believe in this case the documentation is outdated

Ceees2:
@mikepenz 

Took some recordings from the example app. In the first example you can see the arrow at the top right already being rotated up, while the animation is still running. In the second example this is not the case. Since this animation is pretty fast it's barely noticable (I slowed it down in the recording). If there are slower animations this is more present.

With notifyItemChanged = true
before

With notifyItemChanged = false
after

In `SimpleSubExpandableItem.onBind()` the rotation is set:
```kotlin
if (isExpanded) {
    holder.icon.rotation = 0f
} else {
    holder.icon.rotation = 180f
}
```
With `notifyItemChanged = true` the onBind is being called resulting in the rotation being set while the animation for the rotation is still running.

Hope this is more clear.

mikepenz:
@Ceees2 thanks for the examples. that's interesting. I am still hesitating though as this would mean a behaviour change which may break apps. 


---

Could you please rebase on dev to fix the build issue

Ceees2:
@mikepenz Another solution could be changing `toggleExpandable(pos)`, to `toggleExpandable(pos, false)` at line 173 in the onClick.

Currently "fixed" it in our app by checking if animation is null or not in the `onBind()`.

mikepenz:
I'll add a flag so users can modify that behaviour :)

PR #982, opened 2021-02-28
Correct typos in comments and create a dictionary

Benjiko99:
Corrected typos inside of comments.
Also added custom words into a dictionary so that they're not underlined as typos by the IDE.


.gitignore
```diff
@@ -28,6 +28,7 @@ proguard/
 
 #IntelliJ files
 *.iml
-.idea/
+.idea/*
+!.idea/dictionaries
```
Benjiko99, line 32: The change to `.idea/*` is required to allow negating the rule for sub-directories.

mikepenz (approved):
Thank you so much!

